PR #305, opened 2026-06-05
fix(cli): impact --json always emits valid JSON, even when target not found

Disentinel:
## Problem

`grafema impact <target> --json` is an agent-facing machine contract, but when the target node could not be resolved, it wrote the diagnostic to **stderr** and emitted **nothing on stdout** while exiting **0** — a silent success. Any consumer doing `JSON.parse(stdout)` then gets `Unexpected end of JSON input`.

This is the pre-existing failure behind `impact-class.test.ts:622` ("class impact >= single method impact"), flagged as a follow-up in #303: that test runs `impact function findById --json` where `findById` is a `METHOD` (not found when searched as `FUNCTION`), so the method result was empty stdout and the test's `JSON.parse` threw.

`impact` was the **only** CLI command in this failure class — `who`/`trace` already emit valid JSON on empty/not-found, and `describe`/`explain`/`get` signal not-found via a non-zero `exitWithError`. `impact` alone returned **0 + empty stdout**.

## Reproduce (live, before fix)

Minimal graph (one FUNCTION node) populated via `RFDBServerBackend`, then:

```
$ grafema impact "function existingFn"  --json   # found  → valid JSON, status 0
$ grafema impact "function Nonexistent" --json   # missing → status 0, stdout "", JSON.parse → "Unexpected end of JSON input"
```

Actual before-fix stdout for the missing target: `""` → `JSON.parse(stdout) error: Unexpected end of JSON input`.

## Fix

`packages/cli/src/commands/impact.ts` — in the `--json` not-found branch, emit a JSON object that mirrors the success shape (`directCallers` / `transitiveCallers` / `affectedModules` / `callChains`) with `target: null` and a human-readable `error` string. Plain (non-json) text output is unchanged.

After fix, the missing-target stdout is:

```json
{
  "target": null,
  "error": "No FUNCTION \"Nonexistent\" found",
  "directCallers": 0,
  "transitiveCallers": 0,
  "affectedModules": {},
  "callChains": []
}
```

`target: null` plus a non-empty `error` lets a consumer distinguish "found with 0 callers" from "not found" while reading the count/collection fields unconditionally — consistent with `who`/`trace`.

## Spec

- **Given** an analyzed graph, **when** `impact "<missing>" --json`, **then** stdout is a valid JSON object with `target: null`, zeroed counts, and a non-empty `error` (exit 0).
- **Given** a target that exists, **when** `impact "<target>" --json`, **then** stdout is a valid JSON object with `target` populated (unchanged behavior).

## Test (red → green)

New `packages/cli/test/impact-notfound-json.test.ts`. It populates a minimal graph directly through `RFDBServerBackend` (no language analyzer) so it's fast and self-contained, then asserts the `--json` contract for both not-found and found targets.

- **Red** (unfixed): not-found subtest fails at `JSON.parse` (`Unexpected end of JSON input`); found subtest passes. → `1 fail / 1 pass`.
- **Green** (fixed): `2 pass / 0 fail`.

## Adversarial review

- **A — Correctness:** message is plain string interpolation; `type || 'node'` covers bare-name targets; `JSON.stringify` escapes quotes/Unicode in names. Success path untouched. Non-json text path identical to before (still `console.log("No … found")`).
- **B — Structural:** 17-line localized change in one file; no new coupling; reuses the existing success-payload shape.
- **C — Class-not-instance:** swept all CLI commands. `impact` was the sole "status 0 + `--json` + empty stdout" case; `who`/`trace` already emit JSON on not-found, `describe`/`explain`/`get` use non-zero `exitWithError`. No sibling fix or follow-up needed.

## Verification (actual, this run)

- New test: **red 1/2 → green 2/2** (run via the prebuilt rfdb-server).
- `pnpm typecheck` → exit 0. `pnpm lint` → exit 0. CLI build clean.
- Diff: 1 file changed in `src` (17 +/2 −) plus the new test.

### Environment note on the full gate

`pnpm build` (all packages) and `./scripts/test.sh` need the native toolchain — Rust orchestrator + the Haskell `js-analyzer` — which is **not provisioned in this execution environment** (no GHC/cabal), so the `analyze`-dependent sibling tests (incl. `impact-class.test.ts` itself) can't run here. The new test deliberately avoids `analyze` and exercises the exact `JSON.parse` failure those tests hit, so the fixed path is verified end-to-end via the CLI. CI (with the analyzer built) should additionally green `impact-class.test.ts:622`.

Follows up #303 / REG-543.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Vqymbx5dbxJX5XgShHGj5M)_

### DIFF
--- a/packages/cli/src/commands/impact.ts
+++ b/packages/cli/src/commands/impact.ts
@@ -74,10 +74,25 @@ Examples:
       const target = await findTarget(backend, type, name);
 
       if (!target) {
+        const notFoundMessage = `No ${type || 'node'} "${name}" found`;
         if (options.json) {
-          process.stderr.write(`No ${type || 'node'} "${name}" found\n`);
+          // --json is an agent-facing contract: ALWAYS emit a valid JSON object on
+          // stdout, even when the target is missing. A not-found result is still a
+          // valid answer, not an empty stream + a side-channel stderr line (which
+          // makes a consumer's JSON.parse throw "Unexpected end of JSON input").
+          // The shape mirrors the success payload so callers can read
+          // directCallers/affectedModules unconditionally; target=null plus a
+          // human-readable `error` string signal the miss.
+          console.log(JSON.stringify({
+            target: null,
+            error: notFoundMessage,
+            directCallers: 0,
+            transitiveCallers: 0,
+            affectedModules: {},
+            callChains: [],
+          }, null, 2));
         } else {
-          console.log(`No ${type || 'node'} "${name}" found`);
+          console.log(notFoundMessage);
         }
         return;
       }
--- a/packages/cli/test/impact-notfound-json.test.ts
+++ b/packages/cli/test/impact-notfound-json.test.ts
@@ -1,0 +1,101 @@
+/**
+ * Test for `grafema impact <target> --json` ALWAYS emitting valid JSON on stdout,
+ * including when the target node is not found.
+ *
+ * Bug (pre-existing, surfaced by impact-class.test.ts:622 "class impact >= single
+ * method impact"): when `--json` was set and `findTarget` returned null, impact.ts
+ * wrote the "No <type> <name> found" message to *stderr* and emitted nothing on
+ * stdout. An agent consuming the documented `--json` contract then does
+ * `JSON.parse(stdout)` on an empty string and gets "Unexpected end of JSON input".
+ *
+ * `--json` is an AI-first machine contract: a not-found result is still a valid
+ * answer and must be a parseable JSON object, not an empty stream + a side-channel
+ * stderr line. This test pins that contract.
+ *
+ * Setup note: unlike the sibling impact tests, this one populates a minimal graph
+ * directly through RFDBServerBackend instead of running `grafema analyze`. The
+ * not-found path is independent of how the graph was built — it only needs a graph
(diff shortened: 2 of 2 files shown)


Disentinel:
Superseded by #304 (merged), which already landed the `impact --json` not-found fix on `main` (`{ target: null, directCallers: 0, ... }`). This branch now conflicts with `main` (CONFLICTING/DIRTY) and its core change is redundant.

Recommend closing — unless it carries extra coverage worth rebasing. Verified: `packages/cli/src/commands/impact.ts` on `main` already emits the zero-impact JSON object on not-found.

Disentinel:
Agreed — closing as superseded by #304. Verified on `origin/main`: `packages/cli/src/commands/impact.ts` already emits the zero-impact JSON object (`target: null`, zeroed counts) on not-found, and `packages/cli/test/impact-json-not-found.test.ts` already pins the same contract (parseable JSON, `target` null, zero impact) — so both the fix and the test here are redundant, and the branch conflicts with `main`.

The only deltas in this PR were dropping the stderr note and adding an `error` string to the JSON; neither is extra coverage worth rebasing, and keeping the human-readable note on stderr (as `main` does) is the better call. Nothing worth salvaging — closing.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #387, opened 2026-06-11
fix(util): anchor grafema:// virtual marker to authority boundary

Disentinel:
## Summary

`parseGrafemaUri` detected the reserved virtual-node marker `_` by scanning the **whole** string for the substring `/_/`. A real file path containing a bare `_` path segment (e.g. `src/_/util.ts`) was therefore misparsed as a *virtual* node, **silently corrupting** the reconstructed semantic ID. `parseSemanticIdV2` carried a second, divergent copy of the same URI-parsing logic with the identical bug.

This is the same latent class as #386 (semantic-ID parsing silently failing on the `grafema://` URI format).

## Spec

- **Invariant restored:** `toCompactSemanticId(toGrafemaUri(id)) === id` for every semantic ID whose file path contains a `_` segment; `parseSemanticIdV2` returns correct `file`/`type`/`name` for such URIs.
- **Given** a node URI `grafema://localhost/grafema/src/_/util.ts#FUNCTION-%3Efoo`
  **When** parsed, **Then** `filePath = src/_/util.ts`, `semanticId = src/_/util.ts->FUNCTION->foo`, `isVirtual = false` (previously: `isVirtual = true`, `semanticId = util.ts#FUNCTION->foo`).
- **Root cause:** the `_` virtual marker is reserved *only* immediately after the authority (`grafema://{authority}/_/...`). Locating it anywhere in the string aliases with legitimate `_` directories.

## Fix

`packages/util/src/core/GrafemaUri.ts` — split off the authority first (2 segments, or 3 for `github.com`/`gitlab.com`/`bitbucket.org`), then treat the remainder as virtual only when it is exactly `_` or starts with `_/`. Standard nodes parse `file#fragment` from the post-authority remainder.

`packages/util/src/core/SemanticId.ts` — `parseSemanticIdV2` now delegates its `grafema://` handling to the canonical `parseGrafemaUri` (single source of truth for authority detection + the `_` marker), removing the divergent buggy copy. MODULE/virtual/singleton/standard shapes are preserved via the existing compact-form recursion.

## Before / After (live `tsx` run against source)

```
# BEFORE
BUG compact="src/_/util.ts->FUNCTION->foo"
      uri=grafema://localhost/grafema/src/_/util.ts#FUNCTION-%3Efoo
      back="util.ts#FUNCTION->foo"  | v2: file=undefined type=undefined name=undefined

# AFTER
OK  rt  "src/_/util.ts->FUNCTION->foo" -> grafema://localhost/grafema/src/_/util.ts#FUNCTION-%3Efoo -> "src/_/util.ts->FUNCTION->foo"
OK  v2  grafema://localhost/grafema/src/_/util.ts#FUNCTION-%3Efoo => src/_/util.ts|FUNCTION|foo
OK  rt  "MODULE#src/_/a.ts" -> grafema://localhost/grafema/src/_/a.ts#MODULE -> "MODULE#src/_/a.ts"
OK  rt  "src/file.ts->CLASS->C" (github.com/owner/repo authority) -> ... -> roundtrips
```

## Tests

New `test/unit/GrafemaUriUnderscorePath.test.js` (12 cases) — placed top-level because the CI `test:coverage` glob `test/unit/*.test.js` is **non-recursive** and does not pick up `test/unit/core/`.

- Red before fix: 7/8 new structural cases failed for the right reason (the genuine-virtual case already passed).
- After fix:
  - New file: **12/12 pass**.
  - `test/unit/core/GrafemaUri.test.js`: 37/37 pass (existing, unchanged).
  - `FederatedRouterFilePath` / `FederatedRouterSymbolId` / `SemanticAddressResolver` / `SemanticIdV2NestedParent`: all pass.
  - Full CI-gated root glob: baseline **489 pass / 107 fail** → with this change **501 pass / 107 fail** (+12 new passing, **0 new failures**). The 107 pre-existing failures are environmental (native `rfdb-server` binary not built; `mcp`/`cli` `dist` not built) and are identical with the change stashed.
- `eslint` on changed files: clean (exit 0).

## Adversarial review

- **A — Correctness:** multiple `_` segments, `_` under `github.com` 3-segment authority, MODULE nodes whose file has a `_` segment, bracket/counter suffixes, singletons, external-module virtual nodes — all round-trip (live harness above).
- **B — Structural:** removed a duplicated URI parser; `GrafemaUri.ts` now the single source of truth. Import direction is one-way (`SemanticId → GrafemaUri`, no cycle — `GrafemaUri` imports nothing). Both files < 500 lines.
- **C — Class/invariant:** grep confirms exactly two `indexOf('/_/')` sites (`GrafemaUri.ts`, `SemanticId.ts`); both now go through the one corrected implementation.

## Known limitation (pre-existing, by design — not a regression)

A directory literally named `_` at the **repository root** (`_/util.ts`) remains inherently ambiguous with the `authority/_/` virtual marker. Disambiguating it would require changing the URI wire format shared with the Rust orchestrator's `to_uri_format`, which is out of scope. The realistic case (a `_` segment anywhere below the top level, e.g. `src/_/`) is fully fixed.

## Blast radius

Callers of `parseGrafemaUri` / `parseSemanticIdV2` / `toCompactSemanticId` (FederatedRouter cross-repo routing, CLI `query`/`trace`, MCP handlers). Behavior changes **only** for inputs containing a `_` path segment, which previously produced corrupted output — no previously-correct result changes (verified by the unchanged 107/501 split and passing federation/semantic test suites).

---
_Generated by [Claude Code](https://claude.ai/code/session_01FugWRbm5EfGy8FuARYa2G3)_

### DIFF
--- a/packages/util/src/core/GrafemaUri.ts
+++ b/packages/util/src/core/GrafemaUri.ts
@@ -113,20 +113,28 @@ export function parseGrafemaUri(uri: string): ParsedGrafemaUri | null {
 
   const rest = uri.slice(GRAFEMA_SCHEME.length); // "authority/file#fragment"
 
-  // Find the authority: everything up to the first '/' after the host
-  // Authority format: "host/project" (e.g., "github.com/owner/repo" or "localhost/project")
-  // We need to find where authority ends and file path begins.
-  // Convention: authority is "host/path" where host has no '/', so we split on first '/' after host.
-  // Actually, authority can be multi-segment (github.com/owner/repo = 3 segments).
-  // Strategy: find '#' first (fragment delimiter), then split the path part.
+  // Determine the authority boundary first. The authority is "host/project"
+  // (2 segments, e.g. "localhost/grafema") or "host/owner/repo" (3 segments)
+  // for known forge hosts. Everything after the authority is either the
+  // reserved virtual marker "_/..." or a "file/path#fragment".
+  const segments = rest.split('/');
+  const host = segments[0];
+  const authoritySegments =
+    host === 'github.com' || host === 'gitlab.com' || host === 'bitbucket.org'
+      ? 3 // host/owner/repo
+      : 2; // host/project (localhost/grafema)
 
-  const hashPos = rest.indexOf('#');
-  const slashUnderscoreSlash = rest.indexOf('/_/');
+  if (segments.length < authoritySegments) return null;
 
-  if (slashUnderscoreSlash !== -1 && (hashPos === -1 || slashUnderscoreSlash < hashPos)) {
-    // Virtual node: authority/_/encoded_id
-    const authority = rest.slice(0, slashUnderscoreSlash);
-    const encodedId = rest.slice(slashUnderscoreSlash + 3); // after "/_/"
+  const authority = segments.slice(0, authoritySegments).join('/');
+  const afterAuthority = rest.slice(authority.length + 1); // skip the '/' after authority
+
+  // Virtual node: the segment immediately after the authority is exactly "_".
+  // This MUST be anchored to the authority boundary — a "_" path segment deeper
+  // in a real file path (e.g. src/_/util.ts) is NOT a virtual marker, so we
+  // cannot scan the whole string for the substring "/_/".
+  if (afterAuthority === '_' || afterAuthority.startsWith('_/')) {
+    const encodedId = afterAuthority.slice(2); // after "_/"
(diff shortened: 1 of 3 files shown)


Disentinel:
🚫 **QA agent REJECT** (base=main) — held for owner.

```
/opt/launch-ops/qa/qa-review.sh: line 18: /usr/bin/timeout: Argument list too long
```
_autonomous QA gate_

PR #393, opened 2026-06-12
fix(type-inference): streaming + flush — plugin no longer exits 1; plugin failures abort analyze

Disentinel:
## The sole INSTANCE_OF producer was failing on every fresh analyze

Conveyor-diagnosed root cause: the plugin never calls `hello()`, so `queryNodes` falls back to the legacy non-streaming path with a 60s hard timeout — Phase 4's 92,736 CALL nodes die every run (`exit 1` after ~102s). A second latent bug: no `flush()` — under MVCC (visibility=publish) its edges published only as a side effect of later orchestrator commits.

**Fix (+9 lines in the plugin):** `hello()` after connect (negotiates streaming: same query 11s vs 60s-timeout) + `flush()` before close.

**Verified on a dogfood graph copy:** exit 0 in 2m26s; INSTANCE_OF **97 → 2,289** (arithmetic exact); idempotent second run (0 new edges).

## Plugin failures now abort the run

Since 0.4.0 removed the legacy resolve fallback, a silently failing plugin means a silently half-populated graph. `check_plugin_result` in the orchestrator DAG runner aborts on plugin exit/timeout/invalid-output/commit failure (3 unit tests + e2e proven both directions: failing plugin → analyze exit 1; passing → exit 0).

## Known follow-up (found by the adversarial review, pre-existing)

`createBuiltinNodes` read-after-add is invisible under MVCC → builtin `HAS_METHOD` edges orphaned (proven: +215 METHOD nodes, HAS_METHOD unchanged). Tracked separately.

### DIFF
--- a/packages/grafema-orchestrator/src/main.rs
+++ b/packages/grafema-orchestrator/src/main.rs
@@ -1989,7 +1989,9 @@ async fn main() -> Result<()> {
                     }
                 };
 
-                let plugin_results = plugin::run_plugins_dag(
+                // A plugin failure aborts the analyze run (loud-failure policy,
+                // same as derive-pack failures) — but shut the pool down first.
+                let dag_result = plugin::run_plugins_dag(
                     &user_plugins,
                     &mut rfdb,
                     &socket_path,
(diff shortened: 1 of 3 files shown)


Disentinel:
Superseded by #403 — integrated into one verified-together branch (these 6 conflicted pairwise on the shared pack registries; #403 resolves the union with a coherent producer-before-consumer order and runs the full suite + Gate A green together).

PR #396, opened 2026-06-12
fix(js-analyzer): export every declarator of a multi-declarator export (#394)

Disentinel:
Closes #394.

## Spec

**Invariant restored:** an `ExportNamedDeclaration` whose declaration is a multi-declarator `VariableDeclaration` exports *every* binding, not just the first.

**Given** a module `export const a = 1, b = 2;`
**When** the file is analyzed
**Then** the `EXPORT` node has `EXPORTS` edges to **both** the `a` and `b` `CONSTANT` nodes, and the file export index contains both `a` and `b` as `NamedExport` (so `import { b } from './x'` resolves).

**Entities:** `EXPORT` node, `CONSTANT`/`VARIABLE` nodes, `EXPORTS` edges, export-index (`emitExport`).

## Root cause (HEAD `9ac0681`)

`ruleVariableDeclaration` collapsed all declarators to a single `firstId`, and `ruleExportNamedDeclaration`'s declaration branch used that one id + first-declarator `getDeclName` to emit exactly one `EXPORTS` edge + one `emitExport`. Subsequent bindings were created as nodes (correctly `exported=true`) but orphaned from the export surface — invisible to cross-module import resolution. (Originally surfaced by live probing, recorded in `_ai/gaps.md`.)

## Fix (minimal, single source of truth)

- Factor the declarator fold into `ruleVariableDeclarationBindings :: ASTNode -> Analyzer [(Text, Text)]`, returning `(bindingName, nodeId)` for every top-level simple-identifier declarator, in source order, with scope threading preserved (later initializers still resolve earlier bindings).
- `ruleVariableDeclaration` becomes a thin wrapper returning the first binding's id — `walkNode`'s `Maybe Text` contract is unchanged.
- `ruleExportNamedDeclaration` special-cases a `VariableDeclaration` declaration and emits an `EXPORTS` edge + export-index entry **per binding**.

This removes the duplicated declarator-iteration that *was* the bug class.

## Before / after (evidence)

Env: GHC 9.4.7 + cabal 3.8.1.0 (apt), in-process Hspec harness (`analyzeAST` inspects graph shape directly — no orchestrator/RFDB).

**Before** (new test on HEAD — red for the right reason; `a` passes, `b` fails):
```
Exports
  exports every declarator of a multi-declarator export [x]
test/Spec.hs:238: 1) ... expected: True but got: False   -- hasExport "b" NamedExport
```

**After** (`cabal test spec`):
```
Exports
  exports every declarator of a multi-declarator export [v]
[...]

Issue #394, opened 2026-06-12
JS analyzer: `export const a = 1, b = 2` exports only the first declarator (declarators 2+ silently dropped)


Disentinel:
## Summary

A single `ExportNamedDeclaration` whose `declaration` is a multi-declarator `VariableDeclaration` (`export const a = 1, b = 2`) emits the `EXPORTS` edge **and** the export-index entry (`emitExport`) for **only the first declarator** (`a`). Every subsequent binding (`b`, …) is created as a node (correctly `exported=true`) but has **no `EXPORTS` edge from the EXPORT node** and **no entry in the file's export index** — so it is invisible to cross-module import resolution. A consumer doing `import { b } from './x'` will not resolve `b`.

This is one of the two JS-analyzer producer bugs recorded in `_ai/gaps.md` ("Two JS-analyzer producer bugs found by Wave-0 live probing (2026-06-10, wf_07c95bfd-e52)"); the gap note was a live-probe finding. This issue adds the HEAD-verified root cause + a ready-to-apply fix and failing test.

## Root cause (HEAD `9ac0681`)

`ruleVariableDeclaration` collapses every declarator to a single `firstId`:

`packages/js-analyzer/src/Rules/Declarations.hs:50`
```haskell
goDeclarators _ firstId [] = return firstId   -- :: Analyzer (Maybe Text) — only ONE id
```

`ruleExportNamedDeclaration` walks the declaration and gets back only that one id, so it emits exactly one `EXPORTS` edge + one `emitExport`:

`packages/js-analyzer/src/Rules/Declarations.hs:629-640`
```haskell
    Just decl -> do
      mChildId <- withExported $ withAncestor node (walkNode decl)   -- Maybe Text: first declarator only
      forM_ mChildId $ \childId -> do
        emitEdge GraphEdge { geSource = nodeId, geTarget = childId
                           , geType = "EXPORTS", geMetadata = Map.empty }
        let exportName = getDeclName decl                            -- also first-declarator name
        emitExport ExportInfo { eiName = exportName, eiNodeId = childId
                              , eiKind = NamedExport, eiSource = Nothing }
    Nothing -> return ()
```

`getDeclName` (`Declarations.hs:848-858`) likewise returns only the first declarator's name for a `VariableDeclaration`. So both the edge and the index entry are first-declarator-only.

## Expected vs actual

[...]

PR #399, opened 2026-06-12
fix(js-analyzer): index every binding of a destructuring export

Disentinel:
## Spec

**Invariant restored:** a top-level `export const { a, b } = obj` / `export const [a, b] = obj` puts **every** destructured binding into the file export index (`NamedExport`), so a consumer's `import { a } from './x'` resolves.

**Given** a module `export const { a, b } = obj;`
**When** the file is analyzed
**Then** the file export index contains both `a` and `b` as `NamedExport`, matching the `CONSTANT` binding nodes (which already carry `gnExported=True`).

**Entities:** export-index (`emitExport` → `FileAnalysis.faExports`), destructured `CONSTANT`/`VARIABLE` binding nodes.

This is the **destructuring-export gap** flagged as out-of-scope in #394 / PR #396 ("destructuring binding-name remains a pre-existing gap") and recorded in `_ai/gaps.md`.

## Root cause (HEAD `a4d891f`)

`handleDestructuringDecl` (`packages/js-analyzer/src/Rules/Declarations.hs:188`) creates one binding node per destructured name via `extractBindingInfo` and stamps `gnExported=isExported` — but it **never calls `emitExport`**. The export-index entry is emitted once, back in `ruleExportNamedDeclaration` (`Declarations.hs:640-643`), as `emitExport { eiName = getDeclName decl, … }` — and `getDeclName` returns `""` for a pattern (`Declarations.hs:856-866`). So destructured bindings are created as exported nodes yet are **absent from the export surface** — invisible to cross-module import resolution.

## Fix (minimal, co-located, non-overlapping)

In `handleDestructuringDecl`'s per-binding emitter (`emitBindings`), when the declaration is exported (`exported'` — set only under `withExported`, i.e. a top-level `export`), emit a `NamedExport` for each binding's **local** name (`+14` lines):

```haskell
if exported'
  then emitExport ExportInfo { eiName = name, eiNodeId = nodeId, eiKind = NamedExport, eiSource = Nothing }
  else return ()
```

`extractBindingInfo` already yields the importable local name for object shorthand (`{ a }`), array elements (`[a, b]`), renamed props (`{ e: f }` → `f`), defaults (`{ a = 1 }` → `a`) and rest (`{ ...rest }` → `rest`), dropping empties — so the index now matches the binding nodes for the full pattern grammar.

**Why this site, not `ruleExportNamedDeclaration`:** that function + `ruleVariableDeclaration` are being rewritten by in-flight **PR #396** (the multi-declarator export fix, #394). Fixing destructuring at the node-creation site keeps this PR **disjoint** from #396 — no merge conflict, and the two compose cleanly (post-#396 the simple-identifier path skips `name==""`, so no double-emit for destructuring).

## Before / after (actual `cabal test spec`, GHC 9.4.7 / cabal 3.8.1.0)

```
# BEFORE (red, for the right reason — binding nodes exist, index entries don't)
Exports
  exports every binding of an object destructuring export [x]
  exports array and renamed destructuring export bindings [x]
  test/Spec.hs:294: expected: True but got: False   -- hasExport "a" NamedExport
29 examples, 3 failures

# AFTER
Exports
  exports every binding of an object destructuring export [v]
  exports array and renamed destructuring export bindings [v]
29 examples, 1 failure
```

The single remaining failure is **pre-existing and unrelated** — `TypeScript Interfaces / METHOD_SIGNATURE … HAS_PROPERTY` (`Spec.hs:543`), which failed identically in the **red** run when I had made **zero** source changes (same case noted as pre-existing in PR #396). `cabal build grafema-analyzer` links clean under `-Wall -Werror -O2`.

## Tests

`packages/js-analyzer/test/Spec.hs` — two cases in `exportTests` asserting the export index contains every binding:
- object shorthand `export const { a, b } = obj` → exports `a`, `b`;
- array `export const [c, d] = arr` → exports `c`, `d`; renamed `export const { e: f } = obj` → exports the **local** `f` and **not** the key `e`.

The in-process Hspec harness (`analyzeAST`) inspects `FileAnalysis.faExports` directly — the analyzer's real output, no orchestrator needed.

## Adversarial review

- **A — Correctness:** empty pattern `{}` → `extractBindingInfo`=[] → no entry (no junk); non-exported `const {a}=obj` (function-local or top-level non-export) → `exported'`=False → skipped (full suite's existing destructuring tests unregressed); nested `{ a: { b } }` → exports `b`; default `{ a = 1 }` → `a`; rest `{ ...rest }` → `rest`; renamed `{ e: f }` → `f` not `e` (asserted); array holes `[, a]` → null element dropped by `getChildren`, `a` exported. `export let/var` destructuring uses the same path → also indexed.
- **B — Structural:** the export emission is co-located with the binding-node emission in `handleDestructuringDecl` (single source of truth for "what does this pattern bind") — no duplicated pattern-walking. No new imports (`emitExport`/`ExportInfo`/`NamedExport` already used in the module). `Declarations.hs` is a **pre-existing** god-file (>500 lines, flagged in #396); this `+14`-line change does not create that condition and splitting it is out of scope.
- **C — Class / siblings:** `export default` cannot destructure-declare (takes an expression) → `ruleExportDefaultDeclaration` unaffected. CommonJS `module.exports = {…}` is a different mechanism (`Rules/Statements.hs`), not this class. The per-binding **EXPORTS graph edge** for destructured bindings 2+ still lives in `ruleExportNamedDeclaration` (only `firstId` gets it today) — left as a follow-up to avoid overlapping #396; the export **index** (what import resolution consumes, per #394) is the load-bearing fix and is complete here.

## Blast radius

`handleDestructuringDecl` is reached only from `ruleVariableDeclarator` (`Declarations.hs:83-84`) for `ObjectPattern`/`ArrayPattern` ids. The change only **adds** export-index entries, and only when `exported'` is True (top-level `export`). No previously-correct output changes. Pre-`#396`, a harmless `eiName=""` entry from `ruleExportNamedDeclaration` still co-exists (pre-existing, removed when #396 lands); my entries use the real binding names, so there is no collision.

## Follow-ups (not in this PR)

- Per-binding `EXPORTS` graph edge for destructured bindings 2+ (currently only the first binding gets the edge) — owned by `ruleExportNamedDeclaration`, deferred to avoid conflict with in-flight #396.

🤖 https://claude.ai/code/session_01NUamapMGRZJoQqVUAubtr6

---
_Generated by [Claude Code](https://claude.ai/code/session_01NUamapMGRZJoQqVUAubtr6)_

### DIFF
--- a/packages/js-analyzer/src/Rules/Declarations.hs
+++ b/packages/js-analyzer/src/Rules/Declarations.hs
@@ -248,6 +248,20 @@ handleDestructuringDecl patternNode declaratorNode _parentDecl nodeType kind dk 
               , geMetadata = Map.singleton "source" (MetaText "default")
               }
         Nothing -> return ()
+      -- Destructuring under `export const { a, b } = …` / `export const [a, b] = …`:
+      -- each binding must enter the file export index so a consumer's
+      -- `import { a } from './x'` resolves. The simple-identifier export path emits
+      -- this in ruleExportNamedDeclaration via getDeclName, which returns "" for a
+      -- pattern, so destructured bindings would otherwise be absent from the export
+      -- surface despite gnExported=True. `exported'` is True only under withExported
+      -- (a top-level `export`), so non-exported destructuring is unaffected. `name`
+      -- is the LOCAL binding (non-empty — extractBindingInfo drops empties), i.e. the
+      -- importable name (`{ e: f }` exports `f`, not the key `e`).
+      if exported'
+        then emitExport ExportInfo
+               { eiName = name, eiNodeId = nodeId
+               , eiKind = NamedExport, eiSource = Nothing }
+        else return ()
       let firstId' = case firstId of
             Nothing -> Just nodeId
             _       -> firstId
--- a/packages/js-analyzer/test/Spec.hs
+++ b/packages/js-analyzer/test/Spec.hs
@@ -265,6 +265,73 @@ exportTests = do
     let exportIds = nub (map gnId (findNodes "EXPORT" "*:./utils" fa))
     length exportIds `shouldBe` 2
 
+  -- Test 8d: `export const { a, b } = obj` must put EVERY destructured binding in
+  -- the file export index, so a consumer `import { a } from './x'` resolves. On
+  -- HEAD the binding NODES are created (gnExported=True) by handleDestructuringDecl,
+  -- but no emitExport runs for them — ruleExportNamedDeclaration emits a single
+  -- entry whose name is getDeclName(pattern) = "" — so neither `a` nor `b` reaches
+  -- the export surface. (Recorded in _ai/gaps.md as the destructuring-export gap,
+  -- sibling of #394.)
+  it "exports every binding of an object destructuring export" $ do
+    let fa = analyzeAST $ BL.concat
+          [ "{\"type\":\"Program\",\"start\":0,\"end\":40,\"body\":["
+          , "{\"type\":\"ExportNamedDeclaration\",\"start\":0,\"end\":40,"
+          , "\"declaration\":{\"type\":\"VariableDeclaration\",\"start\":7,\"end\":39,\"kind\":\"const\",\"declarations\":["
+          , "{\"type\":\"VariableDeclarator\",\"start\":13,\"end\":38,"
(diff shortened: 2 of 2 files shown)


Disentinel:
✅ **QA gate PASS (review-only)** — base `main` is tier-3, NOT auto-merged. Owner decides the merge. _autonomous QA gate_

PR #424, opened 2026-06-14
fix(rfdb/materialize): reject all-decimal semantic-ids in @materialize_node (E-MAT-010)

Disentinel:
## Source

`_ai/gaps.md` → "@materialize_node advisory follow-ups (W4 review wf_4abcf48c-1d8, 2026-06-10)" item **#1** — *"All-numeric semantic-id drift … extend E-MAT-010 to reject all-decimal sids."* No open GitHub issue / Linear ticket; triaged from the recorded gap.

## SPEC

**Invariant restored:** every semantic-id that `plan_node_writeback` *accepts* must derive the **same** u128 node id regardless of which id-derivation path resolves it later.

The two paths diverge on all-decimal strings:
- **mint** (`derive/materialize.rs:334`): `string_id_to_u128(sid)` → **always BLAKE3**(sid)[0..16].
- **wire resolve** (`bin/rfdb_server.rs:1009` `string_to_id`, `:1022` `resolve_node_id`): `s.parse::<u128>()` **first**, hash only on parse failure.

So a node materialized with sid `"12345"` is minted with id `223388589721311106277013205299978891910` (= blake3) but is looked up as `12345` on every wire round-trip → **silently unreachable by its own sid**.

**Given** a `@materialize_node` predicate whose semantic-id column is an all-decimal string (`"12345"`, `"0"`, `"007"`)
**When** `plan_node_writeback` runs
**Then** it aborts **before any commit** with code `E-MAT-010` (the same abort-no-commit discipline as the other malformed-head cases) — instead of minting an unreachable node.

## Fix

One guard arm in the existing `E-MAT-010` match (`derive/materialize.rs`):

```rust
Value::Str(s) if s.parse::<u128>().is_ok() => { /* E-MAT-010 */ }
```

`parse::<u128>().is_ok()` is the **exact inverse** of the wire path's parse-first branch, so the rejected set is *provably the exact divergence set*. Strings that overflow u128, are negative, contain unicode digits, or are prefixed (the shipped convention) all fail to parse and are hashed by **both** paths → consistent → still accepted.

## Before / After (live)

```
# BEFORE (red) — all-decimal sid is minted, not rejected:
plan_node_writeback(... sid="12345" ...)
  => NodeRecordV2 { semantic_id: "12345", id: 223388589721311106277013205299978891910, ... }
  (blake3("12345"); wire path would resolve 12345 — divergent)

# AFTER (green):
cargo test --profile fast --lib derive::materialize
  => 15 passed; 0 failed
     incl. plan_node_writeback_rejects_all_decimal_semantic_id
```

Test asserts the divergence (`blake3("12345") != 12345`), rejects `"12345"/"0"/"007"` with `E-MAT-010`, and guards the boundary: prefixed `"issue::shape-violation::42"` still mints exactly one node (no regression for real packs).

## Adversarial review

- **Round A (correctness):** guard uses the *same* `parse::<u128>` predicate as the wire path → whatever quirks it has (leading `+`, leading zeros, no whitespace-trim) are identical on both sides; the rejected set is exactly the divergence set. Overflow/negative/unicode-digit/prefixed sids fail to parse → hashed by both → correctly *not* rejected. **Verdict: sound.**
- **Round B (structural):** no new coupling beyond the documented invariant. `string_to_id` lives in the binary crate (`bin/rfdb_server.rs`), not the lib, so a shared helper isn't cleanly reachable — the comment documents the coupling explicitly. *Pre-existing note:* `materialize.rs` is already a >500-line file (mostly tests); extraction is out of scope here.
- **Round C (class-not-instance):** the **edge** path (`plan_writeback`/`value_to_id`, `E-MAT-003`) *intentionally* accepts decimal strings — edge endpoints **reference** existing node ids (the decimal *is* the id), they don't **mint**. So the asymmetry is correct and there's no sibling bug. `plan_node_writeback` is the only place a user-authored `.dl` rule controls a raw minted sid. **No follow-ups filed.**

## Blast radius

Callers: `graph/engine_v2.rs:660`, `:1171` — both propagate via `?` (abort-no-commit). No caller depends on all-decimal sids. Shipped stdlib packs use prefixed sids (`issue::…`, `MODULE#…`), so production output is unchanged.

## Verification

- `cargo test --profile fast --lib derive::materialize` → 15/0
- `cargo test --profile fast --lib` (full rfdb crate) → **1401 passed, 0 failed**, 25 ignored
- `cargo clippy --profile fast --lib` → exit 0, no new warnings (116 pre-existing crate-wide warnings unrelated to this change)

Change is Rust-only in `rfdb-server`, touches no JS/TS — the `pnpm` gate is unaffected.

## Memory write-back

`_ai/gaps.md` item #1 marked ✅ RESOLVED with the evidence above.

---
_Generated by [Claude Code](https://claude.ai/code/session_01UzwNiYRURcmsH67acQHTxG)_

### DIFF
--- a/_ai/gaps.md
+++ b/_ai/gaps.md
@@ -328,6 +328,15 @@ id is identical across generations are not tombstoned by changed-file deletion).
 1. **All-numeric semantic-id drift**: plan_node_writeback blake3-hashes unconditionally;
    the wire writer string_to_id() parses decimal FIRST — a bare-decimal sid mints a divergent
    u128. Unreachable by shipped packs (prefixed sids); extend E-MAT-010 to reject all-decimal sids.
+   **✅ RESOLVED 2026-06-14 (branch `claude/vigilant-lamport-y65ysu`)**: `plan_node_writeback`
+   (`packages/rfdb-server/src/derive/materialize.rs`) now aborts with `E-MAT-010` when the
+   semantic-id column is an all-decimal string — guarded by `s.parse::<u128>().is_ok()`, the
+   EXACT same predicate the wire path (`string_to_id`/`resolve_node_id`, `bin/rfdb_server.rs:1009,1022`)
+   uses, so the rejected set is provably the exact divergence set (and overflowing/negative/prefixed
+   sids that BOTH paths hash stay accepted). Locked by
+   `derive::materialize::tests::plan_node_writeback_rejects_all_decimal_semantic_id` (asserts
+   `blake3("12345") != 12345`, rejects "12345"/"0"/"007", and that prefixed "issue::…::42" still
+   mints one node). Full rfdb lib 1401/0; clippy clean.
 2. **Auto-flush can tear run isolation under buffer pressure**: add_nodes/add_edges end with
    maybe_auto_flush which publishes WITHOUT the run's pending tombstones — a mid-delta flush
    shows adds before retractions (self-heals at the explicit flush). Suppress auto-flush during
--- a/packages/rfdb-server/src/derive/materialize.rs
+++ b/packages/rfdb-server/src/derive/materialize.rs
@@ -287,8 +287,10 @@ pub fn collect_materialize_node_specs(
 ///
 /// Coded aborts BEFORE any commit (I5): a head whose arity is not `3 + len(meta)`
 /// (`E-MAT-009`, the node mirror of `E-MAT-002`), and a semantic-id column that is not a
-/// non-empty string (`E-MAT-010` — column 0 must be the rule-constructed semantic id;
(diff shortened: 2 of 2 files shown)


Disentinel:

```
packages/rfdb-server/src/derive/materialize.rs
```
_QA gate_

Disentinel:
🚫 **QA precheck REJECT** (attempt 2, base=main).

```
packages/rfdb-server/src/derive/materialize.rs
```
_QA gate_

Disentinel:
🚫 **QA precheck REJECT** (attempt 3, base=main).

```
packages/rfdb-server/src/derive/materialize.rs
```
_QA gate_

PR #425, opened 2026-06-14
fix(rfdb/materialize): rewrite owned @materialize_node nodes on surface change (W4 #5)

Disentinel:
## Source

`_ai/gaps.md` → "@materialize_node advisory follow-ups (W4 review wf_4abcf48c-1d8, 2026-06-10)" item **#5** — *"Never-rewrite staleness for OWNED nodes: a re-derived owned node with a CHANGED surface (e.g. ISSUE message after method rename on the same CALL) keeps the stale payload forever — diverges from the plugin's last-write-wins."* No open GitHub issue / Linear ticket; triaged from the recorded gap (sibling of the just-shipped #1 → PR #424).

## SPEC

**Invariant restored:** an `@materialize_node` rule's OWN node is last-write-wins — re-deriving it with a changed surface refreshes it, like a plugin overwriting its own node. (A FOREIGN producer's node at the same id is still never clobbered.)

**Given** an OWNED `@materialize_node` node already stored at id `X` (its `_source` == this rule's hash)
**When** the rule re-derives id `X` with a CHANGED user-visible surface (`name`/`file`/`node_type`/`meta(...)`)
**Then** the stored node is rewritten to the fresh surface — instead of keeping the stale payload forever.

**Idempotency preserved:** a re-run with an UNCHANGED surface writes nothing (the volatile `_generation` field is excluded from the change check). A FOREIGN node at the same id is untouched.

## Root cause (verified at HEAD before fix)

`materialize_writeback_delta` (`packages/rfdb-server/src/graph/engine_v2.rs`) built the node-add set by dropping **every** planned node whose id already exists:

```rust
.filter(|n| !self.store.node_exists_at(snapshot, n.id))
```

and the retraction loop skips any re-derived id (`new_node_ids.contains(&n.id)`). So an owned node re-derived with a changed surface was **neither re-added nor tombstoned → stale forever.**

### Why it's CROSS-FILE (and why the first repro attempt passed)

The bug only bites when the materialized node's `file` is **not** in `changed_files` but its surface depends on a node in a file that **is** — e.g. an ISSUE anchored to the caller's file whose `msg` is a renamed callee in another file. When the node shares the changed file, file-level cleanup (`changed_files` delete-then-insert) sweeps the stale node first and the staleness never surfaces. My initial same-file test passed for exactly this reason; the committed test is deliberately cross-file.

## Fix

An existing node is upserted iff it is OWNED by the same rule **and** its surface differs (`owned_node_surface_changed`):

```rust
.filter(|n| match self.store.get_node_at(snapshot, n.id) {
    None => true,                                   // genuinely new
    Some(stored) => owned_node_surface_changed(n, &stored),
})
```

- **Ownership gate:** rewrite only when `new._source == stored._source` — a foreign producer's node (different/absent `_source`) is never clobbered.
- **Surface check:** `name` / `file` / `node_type` / metadata-minus-`_generation` differs. `_generation` advances every run, so excluding it keeps an unchanged owned node a true no-op.
- **No `delete_node`:** which would cascade-tombstone the node's edges. The upserted new version supersedes the old on read (newest-segment-wins — `get_node_at`/`find_nodes_at` iterate `node_segments.iter().rev()`), so the node's edges are preserved. The rewritten id stays in `new_node_ids`, so the retraction loop never tombstones it.

## Before / After (live)

```
# Cross-file ISSUE: msg = a b.js callee's name; only b.js recommitted on rename.

# BEFORE (red):
materialize_node_rewrites_owned_node_when_surface_changes
  => assertion failed: meta["msg"] left: "v1"  right: "v2"   (stale)

# AFTER (green):
cargo test --profile fast --lib materialize
  => 42 passed; 0 failed; 1 ignored
     incl. materialize_node_rewrites_owned_node_when_surface_changes
```

The test runs three passes: (1) derive ISSUE `msg="v1"`; (2) rename the `b.js` callee with `changed_files=["b.js"]` only, assert the `a.js`-anchored ISSUE survives the commit (the bug surface), re-derive → `msg="v2"` and `(added,removed)==(1,0)`; (3) re-run with no change → `(0,0)`, no new manifest flip (idempotent).

## Adversarial review

- **Round A (correctness):** empty/unparseable metadata → `materialized_source` returns `None` → not eligible → never rewritten (protects foreign/analyzer nodes). Planned nodes always carry `_source` (`plan_node_writeback` stamps provenance). Metadata comparison is order-independent (`serde_json::Map` `PartialEq` is content-based). `_generation`-only differences never trigger a rewrite (locked by run 3). **Verdict: sound.**
- **Round B (structural):** three small documented free helpers (`materialized_source`, `metadata_surface`, `owned_node_surface_changed`); no new coupling. Per-planned-node `get_node_at` is O(new_nodes) point reads (was `node_exists_at`, same order). *Pre-existing note:* `engine_v2.rs` is already a large file — not touched structurally here.
- **Round C (class-not-instance):** the EDGE path (`plan_writeback`) dedups by `(src,dst,edge_type)` only, so an edge whose endpoints are unchanged but whose `meta(...)` columns changed is likewise not refreshed. This is the **pre-existing, module-documented "edge meta-identity caveat"** (`derive/materialize.rs` module docs), a separate design with its own contract — out of scope for this node fix, recorded in `_ai/gaps.md`. Items #2/#3/#4/#6/#7 of the W4 list remain open.

## Blast radius

`materialize_writeback_delta` is shared by all three `@materialize` write paths (`eval_derive_materialize`, `_incremental`, maintain). All existing materialize tests pass unchanged, including the plain-path "an existing id is never re-added" (no surface change → no-op) and the provenance-scoped exclusive-retraction test (foreign node survives). Change is Rust-only in `rfdb-server`; no JS/TS, so the `pnpm` gate is unaffected.

## Verification

- `cargo test --profile fast --lib materialize` → 42/0 (1 ignored)
- `cargo test --profile fast --lib` (full rfdb crate) → **1401 passed, 0 failed**, 25 ignored
- `cargo clippy --profile fast --lib` → exit 0, no new warnings (reworded one doc comment that tripped `doc_list_item_without_indentation`)

## Memory write-back

`_ai/gaps.md` W4 item #5 marked ✅ RESOLVED with the evidence above; the edge-path sibling recorded as a documented caveat.

---
_Generated by [Claude Code](https://claude.ai/code/session_011mrdJyRXFWkDxHjPVmesoR)_

### DIFF
--- a/_ai/gaps.md
+++ b/_ai/gaps.md
@@ -343,6 +343,37 @@ id is identical across generations are not tombstoned by changed-file deletion).
 7. **Cache pinning for always-scratch programs**: node-materializing entries still pin prior
    snapshots in datalog2_materialize_cache for zero benefit; skip the insert.
 
+### ✅ RESOLVED — item #5 (owned-node staleness), 2026-06-14
+
+The `@materialize_node` write-back now REWRITES an owned node whose surface changed since the
+last run (last-write-wins, the plugin contract), instead of keeping the stale payload forever.
+
+- **Root cause (verified at HEAD before fix):** `materialize_writeback_delta`
+  (`packages/rfdb-server/src/graph/engine_v2.rs`) built `added_nodes` by filtering out EVERY
+  planned node whose id already exists (`!self.store.node_exists_at(snapshot, n.id)`), and the
+  retraction loop skipped any re-derived id (`new_node_ids.contains(&n.id)`). So an owned node
+  re-derived with a CHANGED surface was neither re-added nor tombstoned → stale forever.
+- **Why it's CROSS-FILE (and not masked by file-cleanup):** the bug only bites when the
+  materialized node's `file` is NOT in `changed_files` but its surface depends on a node in a
+  file that IS (e.g. an ISSUE anchored to the caller's file whose `msg` is a renamed callee in
+  another file). When the node shares the changed file, file-level cleanup sweeps it first and
+  the staleness never surfaces — which is why the first repro attempt (same-file) passed.
+- **Fix:** an existing node is now upserted iff it is OWNED by the same rule (`_source` matches)
+  AND its user-visible surface differs (`name`/`file`/`node_type`/`meta(...)`, with the volatile
+  `_generation` excluded so unchanged nodes stay a true no-op) — `owned_node_surface_changed`.
+  A FOREIGN producer's node at the same id is still never rewritten. No `delete_node` is used
+  (which would cascade-tombstone the node's edges); the upserted version supersedes the old on
+  read (newest-segment-wins, `get_node_at`/`find_nodes_at` iterate `node_segments.rev()`), so
+  edges are preserved.
+- **Evidence:** RED→GREEN test `materialize_node_rewrites_owned_node_when_surface_changes`
+  (cross-file ISSUE: `msg` "v1"→"v2" after a `b.js`-only rename; run 3 asserts idempotent no-op).
+  Full `rfdb` lib suite **1401 passed / 0 failed / 25 ignored**; `cargo clippy --profile fast
+  --lib` exit 0 with no new warnings. Module doc (`derive/materialize.rs`) updated to match.
+- **Sibling (NOT fixed, documented caveat):** the EDGE path (`plan_writeback`) dedups by
+  `(src,dst,edge_type)` only, so an edge whose endpoints are unchanged but whose `meta(...)`
+  columns changed is likewise not refreshed — this is the pre-existing, module-documented "edge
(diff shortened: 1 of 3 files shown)


Disentinel:

```
packages/rfdb-server/src/derive/materialize.rs
packages/rfdb-server/src/graph/engine_v2.rs
```
_QA gate_

Disentinel:
🚫 **QA precheck REJECT** (attempt 2, base=main).

```
packages/rfdb-server/src/derive/materialize.rs
packages/rfdb-server/src/graph/engine_v2.rs
```
_QA gate_

PR #444, opened 2026-06-15
feat(guarantees): seed `imports-resolve-to-module` Datalog guarantee (REG-1166)

Disentinel:
Closes REG-1166.

## SPEC

**Invariant restored:** every project-internal import resolves to a MODULE. The
v3 GUARANTEE-system replacement for the deleted `BrokenImportValidator` (REG-572)
and the long-disabled `import-has-source` rule.

**Entities:** `IMPORT` node (`name` = module specifier), `IMPORTS_FROM` edge,
`GUARANTEE` node loaded from `.grafema/guarantees.yaml` (the tracked
default-guarantees home), surfaced by `grafema check` / `check_guarantees`.

**Acceptance (BDD):**
- *Given* a relative IMPORT (`./x`, `../y`) with no `IMPORTS_FROM` edge, *when*
  the guarantee runs, *then* it is reported as a violation.
- *Given* an external/bare-specifier IMPORT (`lodash`, `@scope/pkg`, `node:fs`)
  with no `IMPORTS_FROM`, *then* it is **not** reported (no false positive).
- *Given* a relative IMPORT that has an `IMPORTS_FROM` edge, *then* it passes.

**Rule (severity `warning`):**
```datalog
violation(X) :- node(X, "IMPORT"), attr(X, "name", S), starts_with(S, "."),
                \+ edge(X, _, "IMPORTS_FROM").
```

### Honesty scope (the crux of REG-1166)
A blanket `node(X,"IMPORT"), \+ edge(X,_,"IMPORTS_FROM")` reports **every**
external dependency import as a violation, because external packages have no
`MODULE`/`EXTERNAL_MODULE` leaf node yet (REG-624, REG-687 — see
`js_module_imports.dl`: only relative/workspace specifiers resolve). Scoping to
relative specifiers (`starts_with(name, ".")`) yields **zero external-dependency
false positives with no dependency on REG-624/687** — the "scope to
project-internal imports" option the issue offers.

Evidence the scope is sound: the JS analyzer stamps `IMPORT.name` = the module
specifier (`packages/js-analyzer/src/Rules/Declarations.hs`, `gnName = source`);
relative specifiers begin with `.`. This is conservative across languages —
Rust/Haskell/Java internal modules (`std::`, `Data.Map`, `com.foo`) don't start
with `.`, so the rule simply doesn't fire for them (consistent with "wait for
REG-687"). Granularity choice: target `IMPORT` (module-level, matching the
guarantee name "…to-module" and the disabled predecessor), not `IMPORT_BINDING`
(whose `name` is the binding's local name, not the specifier).

## Changes
- **`.grafema/guarantees.yaml`** — replace the disabled `import-has-source` block
  (incl. its `TODO: REG-XXX` and commented-out rule) with the active,
  relative-scoped `imports-resolve-to-module` guarantee + a thorough description
  (the file's canonical per-guarantee doc location).
- **`test/unit/imports-resolve-to-module.test.js`** — new TDD test (7 cases) that
  runs the **actual seeded rule string** from the YAML against synthetic graphs
  (file & test can't drift).
- **`test/unit/orphaned-node-guarantees.test.ts`** — retire the two stale
  `import-has-source` tests (they referenced a commented-out guarantee and used
  `IMPORT.name='foo'`, which never matched the analyzer).

## BEFORE / AFTER evidence

**Red → Green (focused test):**
```
# before seeding the guarantee:  # tests 7  # pass 0  # fail 7
# after  seeding the guarantee:  # tests 7  # pass 7  # fail 0
```

**`grafema check` surfaces it** (via `GuaranteeManager.import` → `list` → `check`,
the exact path `check.ts` uses, loading the real YAML):
```
seeded guarantee present in list(): true
  name    : imports-resolve-to-module
  severity: warning
  rule    : violation(X) :- node(X, "IMPORT"), attr(X, "name", S), starts_with(S, "."), \+ edge(X, _, "IMPORTS_FROM").
check() passed: false | violations: 1   # ./gone flagged; lodash NOT flagged
```

**Full gate (all green):**
```
pnpm build      → exit 0
./scripts/test.sh → # tests 741  # pass 741  # fail 0   (incl. ok 79 REG-1166)
pnpm typecheck  → exit 0
pnpm lint       → exit 0
```

## Adversarial review
- **Round A (correctness):** edge cases reasoned + tested — empty name, `@scope/pkg`,
  `node:fs`, `lodash/fp` excluded; `./`, `../` included; resolved relative passes.
  Live-verified that `attr(X,"name",S)` binds the free var and `starts_with` filters
  (violation count exactly 1 in a mixed graph).
- **Round B (structural):** config + small test; no god-file. The test's one-block
  YAML extractor is a deliberate tripwire (single-line rule) — drift fails loudly,
  not silently.
- **Round C (class/invariant):** the disabled `import-has-source` was the only
  sibling — retired its stale tests. New guarantee passes the auto-discovery
  "parses without error" tripwire in `orphaned-node-guarantees.test.ts`.

## Blast radius
`.grafema/guarantees.yaml` is consumed by `grafema check` (`GuaranteeManager.import`)
and the auto-discovery test. No code paths change; the guarantee is `warning`, and
`grafema check` is **not** a CI gate, so this cannot break CI. If grafema's own graph
contains unresolved relative imports (or relative asset imports — see caveat), they
will now surface as warnings — the intended signal.

## Follow-ups (siblings found, out of scope)
- **`method-has-class` test bug** (pre-existing, unrelated): the fixture adds a
  `HAS_MEMBER` edge but the rule + analyzer use `HAS_METHOD` (`Declarations.hs:449`),
  so `method-has-class: passes when METHOD has HAS_MEMBER` fails on baseline too.
  One-line fixture fix; not touched here to keep scope tight.
- **REG-624 / REG-687**: when `EXTERNAL_MODULE` leaf nodes land, widen this rule (or
  add a dedicated `external-imports-resolve` guarantee) to target external imports too.

https://claude.ai/code/session_01AytKeDZnarrayB1wxdSwH3

---
_Generated by [Claude Code](https://claude.ai/code/session_01AytKeDZnarrayB1wxdSwH3)_

### DIFF
--- a/.grafema/guarantees.yaml
+++ b/.grafema/guarantees.yaml
@@ -312,17 +312,37 @@ guarantees:
   # it can't fulfill its purpose without the defining edge.
   # ============================================================
 
-  # DISABLED: External library imports (npm/Hackage/crates.io) intentionally lack
-  # IMPORTS_FROM edges — no MODULE node exists for external packages.
-  # TODO: Create EXTERNAL_MODULE virtual nodes for external deps (REG-XXX)
-  # - name: import-has-source
-  #   description: >
-  #     Every IMPORT node must have an outgoing IMPORTS_FROM edge
-  #     to its source MODULE. Without it, dependency tracing is blind —
-  #     we know something is imported but not from where.
-  #   check: datalog
-  #   rule: 'violation(X) :- node(X, "IMPORT"), \+ edge(X, _, "IMPORTS_FROM").'
-  #   severity: warning
+  - name: imports-resolve-to-module
+    description: >
+      Every project-internal IMPORT (a relative specifier — one whose module
+      string starts with ".", e.g. "./util" or "../shared/db") must have an
+      outgoing IMPORTS_FROM edge to the MODULE it resolves to. A relative import
+      with no IMPORTS_FROM is a broken import: the target file was renamed,
+      moved, or never existed, and any cross-file call/dataflow tracing through
+      it is silently blind. This is the v3 GUARANTEE-system replacement for the
+      deleted BrokenImportValidator (REG-572) and the long-disabled
+      `import-has-source` rule (REG-1166).
+
+      SCOPE — why relative-only: EXTERNAL dependency imports (npm "lodash",
+      Hackage "Data.Map", crates "serde", builtins "node:fs") legitimately lack
+      IMPORTS_FROM today, because no MODULE/EXTERNAL_MODULE leaf node exists for
+      them yet (REG-624, REG-687). A blanket rule would report every external
+      import as a violation. Relative specifiers are always project-internal and
+      MUST resolve, so scoping to `starts_with(name, ".")` yields zero
+      external-dependency false positives with no dependency on REG-624/687.
+      The JS analyzer stamps IMPORT.name = the module specifier
+      (js-analyzer/src/Rules/Declarations.hs: gnName = source); relative
+      specifiers begin with ".". When EXTERNAL_MODULE leaf nodes land, widen
+      this rule (or add a dedicated `external-imports-resolve` guarantee).
+
+      KNOWN CAVEAT: relative imports of non-code assets ("./styles.css",
+      "./data.json") are also "." specifiers but have no MODULE node, so they
+      surface here too — Grafema does not yet model asset modules. Treat such
+      hits as a modeling gap, not necessarily a broken import; hence `warning`,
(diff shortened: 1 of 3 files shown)


Disentinel:
✅ **QA gate PASS (review-only)** — base `main` tier-3, not auto-merged. Owner decides. _QA gate_

PR #469, opened 2026-06-19
fix(derive): grafema self-analyze at scale — cap-lift + compaction-suppress + plan-guard

Disentinel:
**Goal:** make `grafema analyze` complete on grafema's OWN monorepo (~714k nodes / 1.33M edges). It failed three different ways at self-scale; all three fixed here. **With this branch the full derive phase completes — 40 packs, 0 SIGSEGV, 0 plan rejects.**

## Fixes (all in `packages/rfdb-server`)

1. **cap-lift** — `@materialize` batch path used the interactive `EvalLimits::default()` 100k `max_intermediate_results` (the deadline was already lifted, the cap was overlooked). `js_local_refs` overflows it. Lifted for the batch path (default unbounded, env `RFDB_MATERIALIZE_MAX_INTERMEDIATE`). `bin/rfdb_server.rs`.

2. **compaction vs parallel-derive heap corruption** (the hard one). Materialize holds a `db.engine.read()` lock but commits derived edges across 40 packs; those commits auto-triggered compaction (rayon, memmap2) which rewrites/unmaps segments **while the same materialize's `par_join_rows` rayon tasks read them** → use-after-unmap heap corruption → SIGSEGV. Confirmed: TSan named `join_derived`/`par_join_rows`; `rayon=1` passes; disabling compaction passes. Fix: `AutoCompactionSuppressGuard` suppresses auto-compaction for the materialize phase; deferred compaction runs at the natural barrier (`end_bulk_load`/explicit `Compact`) under the exclusive write lock. **Correct phase serialization, not masking.** `coordinator.rs` + `bin/rfdb_server.rs`.

3. **MAX_MATERIALIZED_FACTS planner guard (E-PLAN-003)** — `static_member` in `js_property_access_full` estimates 11.7M facts > the 10M guard and is rejected, but the **actual output is ~811 edges** (a ~14,000× cardinality q-error: the estimator cross-products relation sizes, ignoring the highly-selective file+class+method join keys). Made the guard env-overridable (`RFDB_MAX_MATERIALIZED_FACTS`, default 10M). `plan.rs`.

## Verify
```
RFDB_MATERIALIZE_DEADLINE_SECS=3600 RFDB_MAX_MATERIALIZED_FACTS=200000000 \
  grafema analyze --quickstart --verbose   # derive completes, 40 packs, 0 SIGSEGV
```

## Open (separate, for follow-up / workers)
- [ ] **Wall 3 — enricher RPC 60s client timeouts** (`packages/rfdb/dist/client.js`) + slow `clear`/drop of a large DB block full exit-0 + `mcp:tool=27`. The derive is done; the post-derive enrichers time out at scale. Raise/scope the RPC timeout; investigate drop perf.
- [ ] **CI self-analyze job** (non-blocking first) so this never regresses silently (its absence hid all of the above).
- [ ] **Estimator q-error** — selectivity-aware cardinality estimation so E-PLAN-003 stops spuriously tripping (then the env override is unnecessary).
- [ ] Self-analyze is also unacceptably **slow** (~14 min); perf pass.

### DIFF
--- a/packages/rfdb-server/src/bin/rfdb_server.rs
+++ b/packages/rfdb-server/src/bin/rfdb_server.rs
@@ -3045,10 +3045,31 @@ fn dispatch_materialize_datalog(
     limits.deadline =
         Some(std::time::Instant::now() + std::time::Duration::from_secs(materialize_deadline_secs));
 
+    // Same E-EXEC-001 reasoning as the deadline above: the 100k `max_intermediate_results` is an
+    // interactive-query bound and is WRONG for a batch full materialize. Intermediate relations of
+    // legitimate stdlib packs (e.g. js_local_refs scope-walk: ref_scope/inner_of/chain_declares)
+    // scale ~linearly with reference count, so a cold materialize over a non-trivial graph routinely
+    // exceeds 100k rows in a single stratum (the ~21k-node mcp graph already produces ~141k). The
+    // batch path is build-step-like; its real guards are the deadline + cancellation + memory, NOT a
+    // fixed row count. Lift it (default: unbounded). Tunable via `RFDB_MATERIALIZE_MAX_INTERMEDIATE`.
+    let materialize_max_intermediate = std::env::var("RFDB_MATERIALIZE_MAX_INTERMEDIATE")
+        .ok()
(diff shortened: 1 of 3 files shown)


Disentinel:

PR #470, opened 2026-06-19
ci: non-blocking grafema self-analyze regression guard

Disentinel:
## Why

Grafema's CI runs CodeQL **Analyze** jobs — but it never runs *grafema's own* `grafema analyze` against this repo. That gap is what let a whole class of self-scale bugs ship silently:

- a derive-phase **SIGSEGV** (compaction vs. parallel-derive heap corruption), and
- several conservative planner/eval caps (`max_intermediate_results`, `MAX_MATERIALIZED_FACTS`)

These only bite when grafema analyzes its **own** ~714k-node / 1.33M-edge monorepo, and are unreachable by the unit/integration tests (which run on tiny fixture graphs). All of the above were fixed on `fix/derive-batch-cap` (#469); this PR adds the guard so they can't regress silently again.

## What this adds

New workflow `.github/workflows/self-analyze.yml`:

1. **Builds grafema as it ships** — `pnpm build` (TS) + `cargo build --release` for the two native binaries the analyze pipeline spawns (`rfdb-server` + `grafema-orchestrator`, auto-detected from `packages/*/target/release/`).
2. **Runs the full self-analyze** — `grafema analyze --clear` on the whole monorepo via the checked-in dogfooding config (`.grafema/config.yaml`), with the self-analyze env from `fix/derive-batch-cap`:
   `RFDB_MATERIALIZE_DEADLINE_SECS=3600 RFDB_MAX_MATERIALIZED_FACTS=200000000`.
3. **Asserts exit 0 AND that the JS enrichers fired** — the graph must contain `mcp:tool` nodes (produced by the behavior enricher on `packages/mcp/src/server.ts`). A clean exit with an empty graph is the silent-failure mode this guard explicitly catches.
4. **Canaries the derive log** for `SIGSEGV` / plan rejects.

## Runner / scope choice

- **Runner:** `ubuntu-latest` (GitHub-hosted, 16 GB RAM / 4 vCPU). A full self-analyze peaks ~7.5 GB RAM / ~14 min — it fits in 16 GB with headroom.
- **Scope:** the **full monorepo** (not a scoped subset), so all ~40 derive packs are exercised — including the ones that triggered the SIGSEGV (`@stdlib/js_local_refs`) and the cap rejects (`js_property_access_full`). A scoped single-package analyze would not reproduce those cross-package derive paths, so scoping was rejected. If a future graph outgrows 16 GB, the header documents the escape hatch (larger runner label, or scope via `--service`).

## Non-blocking (intentional)

The job is **non-blocking**: `continue-on-error: true`, not a required status check. As of `fix/derive-batch-cap` the derive phase completes with 0 SIGSEGV, but a full self-analyze still fails **late** on an enricher RPC 60s client timeout (being fixed separately on `fix/enricher-rpc-timeout`). Until self-analyze is green end-to-end, this job **runs and surfaces** the result without blocking PRs.

There is a `TODO(make-required)` in the workflow header: once `fix/enricher-rpc-timeout` lands and a full self-analyze reaches exit 0 with `mcp:tool > 0` reliably, remove `continue-on-error: true` and add `self-analyze` to branch-protection required checks.

## Validation

- **actionlint v1.7.12** — exit 0, no warnings.
- **Assert logic dry-run on a real graph** (grafema-dev box, 503k-node live graph): `grafema query --type mcp:tool "" --json` exits 0 and returns a parseable JSON array of `mcp:tool` nodes (61 on that graph). Confirms the exit-0 + non-empty-enricher assertion works against a real self-analyze graph.

### DIFF
--- a/.github/workflows/self-analyze.yml
+++ b/.github/workflows/self-analyze.yml
@@ -1,0 +1,177 @@
+# Grafema self-analyze regression guard
+#
+# WHY THIS EXISTS
+# ---------------
+# Grafema's CI runs CodeQL "Analyze" jobs — it does NOT run *Grafema's own*
+# `grafema analyze` against this repo. That gap is exactly what let a whole class
+# of self-scale bugs ship silently: a derive-phase SIGSEGV (compaction vs.
+# parallel-derive heap corruption) plus several conservative planner/eval caps
+# (max_intermediate_results, MAX_MATERIALIZED_FACTS) that only bite when grafema
+# analyzes its OWN ~714k-node / 1.33M-edge monorepo. None of that is reachable by
+# the unit/integration tests, which run on tiny fixture graphs. See
+# fix/derive-batch-cap (Disentinel/grafema#469).
+#
+# WHAT THIS JOB DOES
+# ------------------
+#   1. Builds grafema as it ships: pnpm build (TS) + cargo --release for the two
+#      native binaries the analyze pipeline spawns (rfdb-server + orchestrator).
+#   2. Runs `grafema analyze` against this repo using the checked-in dogfooding
+#      config (.grafema/config.yaml), with the env that makes self-analyze pass.
+#   3. Asserts the run reached exit 0 AND the JS enrichers actually fired — the
+#      graph must contain `mcp:tool` nodes (produced by the behavior enricher on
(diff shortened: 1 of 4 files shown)


Disentinel:

PR #472, opened 2026-06-19
fix(derive): selectivity-aware per-rule estimate for shared-variable equality joins (E-PLAN-003 q-error)

Disentinel:
## Problem

The derive planner's per-rule output estimate (the `E-PLAN-003` / `MAX_MATERIALIZED_FACTS` guard) cross-products every generator's **raw** relation magnitude along the join, ignoring that a later leg applying an **equality selection** on a generator-introduced key against an already-bound value makes that join nearly 1:1.

The stdlib `static_member` rule (`js_property_access_full.dl`):

```
static_member(PA, T) :- uc_pa(PA,F,N,BaseN), node(Cls,"CLASS"),
    attr(Cls,"file",F), attr(Cls,"name",BaseN),
    edge(Cls,T,"HAS_METHOD"), attr(T,"name",N).
```

estimated **~11.7M facts** on the full monorepo self-analyze graph (|uc_pa| × |CLASS| × avg-degree) and was rejected with `E-PLAN-003`, vs an **actual ~811 materialized edges** — a ~14,000× over-estimate. The `file`/`class-name`/`method-name` equality joins are nearly 1:1, but the estimator never accounted for them. The base branch (`fix/derive-batch-cap`, d26a1e55) worked around it with the `RFDB_MAX_MATERIALIZED_FACTS` env override and explicitly flagged "deeper fix = selectivity-aware estimation (follow-up)." **This PR is that follow-up.**

## Fix (principled, not "raise the constant")

When a generator-introduced variable is **equality-pinned** by a downstream base-relation selection whose other positions are all already-bound — a shared-variable join key: `attr(v,key,BoundVal)` / `attr(BoundId,key,v)` / `edge(v,BoundEndpoint,_)` — the generator is in truth a keyed probe, not a full scan. Its multiplicative contribution is reduced by the existing key-narrowing (√) factor per pinned key instead of multiplying the full relation magnitude.

A key with **no** downstream equality is left unreduced, so a **genuine cross-product** (whose legs share no pinning equality) keeps its full magnitude and still trips the guard — the guard's job is intact. The reduction is a pure ordering/sizing change: it can only let more rules pass, never change which facts a rule derives (I1).

The pin walk advances the running bind set across intervening legs, so an equality whose value side is bound by a sibling leg between the generator and the filter is still recognized.

## Verification

**Unit (plan.rs):**
- `selective_equality_join_estimate_is_realistic_static_member_shape` — the static_member-shaped selective join now plans under the **10M default** guard.
- `genuine_cross_join_without_selective_key_still_trips_guard` — a real two-hop explosion (`edge(X,Y,"T"), edge(Y,Z,"T")` on a dense graph) still trips `E-PLAN-003`.
- `downstream_equality_pin_detection_discriminates` — the pin detector fires on a bound-value equality, NOT on a bare type-restriction, a free-endpoint hop, or a free-value read.
- All **13** plan tests + **323** derive-module tests pass.

**End-to-end (real graph, grafema-dev):** ran a real-graph planner probe over the shipped `js_property_access_full` pack against the live 503k-node / 1.1M-edge self-analyze store, **without** `RFDB_MAX_MATERIALIZED_FACTS` — the pack plans (max `static_member` estimate 7.17M ≤ 10M default guard). The env override is no longer required for this pack on that graph.

> Note: the original 11.7M rejection was observed on the larger ~714k-node JS-heavy monorepo graph (not available as a prebuilt artifact on the box); on the available graphs the CLASS arm is already small via the branch's `order_literals` improvements. This fix makes the estimate robust to that ordering regardless — it targets the exact selective-join mechanism the base commit documented.

### DIFF
--- a/packages/rfdb-server/src/bin/rfdb_server.rs
+++ b/packages/rfdb-server/src/bin/rfdb_server.rs
@@ -3045,10 +3045,31 @@ fn dispatch_materialize_datalog(
     limits.deadline =
         Some(std::time::Instant::now() + std::time::Duration::from_secs(materialize_deadline_secs));
 
+    // Same E-EXEC-001 reasoning as the deadline above: the 100k `max_intermediate_results` is an
+    // interactive-query bound and is WRONG for a batch full materialize. Intermediate relations of
+    // legitimate stdlib packs (e.g. js_local_refs scope-walk: ref_scope/inner_of/chain_declares)
+    // scale ~linearly with reference count, so a cold materialize over a non-trivial graph routinely
+    // exceeds 100k rows in a single stratum (the ~21k-node mcp graph already produces ~141k). The
+    // batch path is build-step-like; its real guards are the deadline + cancellation + memory, NOT a
+    // fixed row count. Lift it (default: unbounded). Tunable via `RFDB_MATERIALIZE_MAX_INTERMEDIATE`.
+    let materialize_max_intermediate = std::env::var("RFDB_MATERIALIZE_MAX_INTERMEDIATE")
+        .ok()
+        .and_then(|v| v.parse::<usize>().ok())
+        .unwrap_or(usize::MAX);
+    limits.max_intermediate_results = materialize_max_intermediate;
+
     // Empty source ⇒ the canonical bundled depends.dl; "@stdlib/<name>" ⇒ a named bundled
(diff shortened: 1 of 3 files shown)


Disentinel:

